PR #21, opened 2020-08-27
keras nlp scoping

tanzhenyu:


rfcs/20200826-keras-nlp-scoping-design.md
```diff
@@ -0,0 +1,79 @@
...
+- Tasks that improves model training.
+- Operations that will work in CPU/GPU/TPU.
+
+Boundaries between keras-nlp and TF.Text:
+- TF.Text will contain all pre-processing operations, such as WordPiece Tokenizer, n-grams, that handles strings.
+- keras-nlp will contain components after tokenization.
+
+Boundaries between keras-nlp and Tensorflow Addons:
+- Highly experimental modeling, layers, losses, etc, live in addons.
+- Components from addons will graduate to Model Garden, given it incurs more usage,
```
bhack, line 49: I suppose we graduate on Tensorflow. Right? Or also in Keras when it will be in the standalone repository again?
We already started a discussion at https://github.com/tensorflow/community/pull/241 some months ago about considering Keras standalone as an upstream target as Tensorflow or not. 
If this is the case, like with the next Keras-cv RFC, we need a little bit to disambiguate this for contributors.

tanzhenyu: Graduating to Keras or TF is mostly a matter of "if this is a Keras object or not", right? This is hopefully clear enough

bhack: Yes this was taken for granted what I meant is that could be confusing for contributors to start a PR in Addons or in a "satellite" Keras repository.
Also our experience is that it is hard to enforce to open a ticket before starting to work on a PR but at least with a policy of public issues before implementation we could try to have a better visibility on the third party contributors routing process.

Sometimes PR a re-routed downstream from Tensorflow to TFA. Do you will also re-route downstream from Keras standalone to keras-nlp (or keras-cv), tfa? 

Then I suppose also that here you will have some kind of "internal" development resources/efforts so I think that for these activities the process could be a little bit more under control then "sparse" user contributions that instead they can instead start in every mentioned repository. 
Having a little bit of visibility on this could help us to not spent time to review PRs that have some parallel internal potential duplicated activities.

tanzhenyu:
Closing this, replaced by https://github.com/keras-team/governance/pull/22

PR #22, opened 2020-08-28
keras-nlp scoping RFC.

tanzhenyu:


tanzhenyu:
from @bhack:
"
Yes this was taken for granted what I meant is that could be confusing for contributors to start a PR in Addons or in a "satellite" Keras repository.
Also our experience is that it is hard to enforce to open a ticket before starting to work on a PR but at least with a policy of public issues before implementation we could try to have a better visibility on the third party contributors routing process.

Sometimes PR a re-routed downstream from Tensorflow to TFA. Do you will also re-route downstream from Keras standalone to keras-nlp (or keras-cv), tfa?

Then I suppose also that here you will have some kind of "internal" development resources/efforts so I think that for these activities the process could be a little bit more under control then "sparse" user contributions that instead they can instead start in every mentioned repository.
Having a little bit of visibility on this could help us to not spent time to review PRs that have some parallel internal potential duplicated activities."

rfcs/20200826-keras-nlp-scoping-design.md
```diff
@@ -0,0 +1,108 @@
...
+- Ability for the open-source community to build more advanced solutions *on top* of this package instead of reinventing it
+- Ability for research scientists to benefit from subclassing and customizing base components to quickly test new research ideas
+
+## Design Proposal
+
+`keras-nlp` will include most standard Transformer-based modules, specifically:
+
+- Keras layer components such as Transformer encoder and decoder blocks.
+- Keras task components such as masked language, span labeler and named entity recognition.
+- Tensorflow operations such as beam search.
```
seanpmorgan, line 46: Does this scope include custom op kernels for things such as beam search? [TFA seq2seq uses one](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/seq2seq/cc/kernels/beam_search_ops.cc) but that would seem to differ from the python only Keras repositories that I thought were being created.

rfcs/20200826-keras-nlp-scoping-design.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Boundaries between keras-nlp and tf.text
+
+- `tf.text` will contain all pre-processing operations, such as WordPiece Tokenizer, n-grams, that handles strings.
+- `keras-nlp` will contain modeling components that cover workflows past the tokenization stage.
+
+## Boundaries between `keras-nlp` and TensorFlow Addons:
+
+- Highly experimental modeling, layers, losses, etc, live in Addons (e.g. newly published research code).
+- Components from Addons will graduate to Model Garden, given they get sufficient usage,
```
seanpmorgan, line 66: So this RFC never went through... but it is news to Addons that we will be graduating to Model Garden. Would like to have a meeting to discuss and finalize this.
https://github.com/tensorflow/community/pull/241
cc @karmel

rfcs/20200826-keras-nlp-scoping-design.md
```diff
@@ -0,0 +1,108 @@
...
+# Boundaries between keras-nlp and core Keras
+
+- `keras-nlp` will contain NLP-specific components
+(e.g. the `MultiHeadAttention` layer may be used outside of NLP, and thus is shipping in core Keras).
+- Components from keras-nlp can graduate to Keras core, given its usage expands beyond
+ natural language processing.
+
+## Dependencies
+
+- Tensorflow version >= 2.4
```
seanpmorgan, line 86: Could we elaborate on why 2.4 is the minimum version to utilize this library?

seanpmorgan (commented):
Some questions on the scope, but overall I think this is a nice change for TF/Keras end user experience.

rfcs/20200826-keras-nlp-scoping-design.md
tanzhenyu: Same as Keras core that I would like this to be pure python library, or at least for the current scoping. TF has mechanism to support things that cannot be interpreted as pure numpy ops.
Specifically, for beam search, I wonder if this would be something tf.text can take -- worth discussing between us.

rfcs/20200826-keras-nlp-scoping-design.md
tanzhenyu: +1

rfcs/20200826-keras-nlp-scoping-design.md
tanzhenyu: We need support of several advanced features to make this very readable: 
1) kwargs as call signature, it is currently available in nightly (for sometime), 2.3 might already supported
2) dict inputs and dict outputs & saved model

2.4 might be the first stable release that these are supported, but we can also claim that statement for >= a certain nightly build

rfcs/20200826-keras-nlp-scoping-design.md
```diff
@@ -0,0 +1,108 @@
...
+- Components from Addons will graduate to Model Garden, given they get sufficient usage,
+and given that they work on CPU/GPU/TPU. The API interface will remain experimental for a short time after graduation,
+so as to leave us the option to make changes based on user feedback.
+
+## Boundaries between keras-nlp and Model Garden
+
+- End to end modeling workflow and model specific details live in Model Garden
+- Model garden will re-use most of the building blocks from keras-nlp
+- Components from Model Garden can graduate to keras-nlp, given they get sufficient usage,
+and given that they work on CPU/GPU/TPU. The API interface should remain stable after graduation.
```
seanpmorgan: @tanzhenyu per discussion at the SIG Addons meeting could we update this to say modular components should not be housed in model garden, and that with sufficient usage/reliability Addons components will be graduated to keras/tf-core as needed/

PR #27, opened 2020-11-19
Add RFC for Keras training property

gdj0nes:
Introduce a property for `training` to the keras base layer to better support transfer learning. Initially raised in this [issue](https://github.com/tensorflow/tensorflow/issues/39036)

### DIFF
--- a/rfcs/20201119-keras-training-property.md
+++ b/rfcs/20201119-keras-training-property.md
@@ -1,0 +1,83 @@
+# Title of RFC
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Gareth Jones (gareth@example.org)                    |
+| **Sponsor**   | A N Expert (expert@example.org)                      |
+| **Updated**   | 2020-11-19                                           |
+
+
+## Objective
+
+Improve the ability to control the `training` arg to by introducing `training` property on the base layer. 
+
+## Motivation
+
+It is difficult to fine tune many kinds of models because the training
+argument is difficult to access. Adding a property would make it easier to
+configure more modular models like segmentation models to better support 
+transfer learning applications
+
+```python
+# A keras Model with two sub networks, backbone and head
+model = tf.keras.Model(...)
+# model.backbone
+# model.head
+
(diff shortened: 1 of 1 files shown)


fchollet (commented):
Thanks for the proposal.

Fundamentally, "training/inference" is not a property of a model (i.e. it is not an element of state), it is a property of a specific call of a model. The same model can be called both in training mode or in inference mode.

The `training` property introduced seems to act as an override for the `training` argument in `call`. There are several concerns:

- Overriding the value passed by users in `call` is rarely a good idea. It will be a source of bugs.
- It may be a source of confusion with the `trainable` property.

I don't think adding a `training` property is the solution. I'd like to better understand what motivated you to proposed it:

> It is difficult to fine-tune many kinds of models because the training argument is difficult to access.

What was your use case and why was it difficult?

gdj0nes:
Thanks for reviewing my proposal.

The challenge I am trying to solve is sharing models that can be fine tuned. The idea comes from your [colab](https://colab.research.google.com/drive/17vHSAj7no7RMdJ18MJomTf8twqw1suYC#scrollTo=gWgNGY9S8q6d) on transfer learning in Keras . When fine tuning or reusing existing models, it is challenging to manipulate the execution flow because the training argument for specific sub layers cannot be changed. The lack of control over that execution pathway makes it more difficult to reuse and incorporate models that were not designed explicitly to be extensible.

Concretely, any model that uses the training parameter to alter execution flow, e.g. a two stage detector, will not be able to finetune without also setting batchnorm layers to training. 

A training property would allow a user to selectively manipulate the argument per layer. In the two stage detector example, you would set the training arg to True for the model and iterate over the batchnorm layers in the bbox head, setting training to false on each using the property. I agree that this does complicate things and may contribute to confusion between training and trainable.

Generally speaking, reducing the reliance on the training argument to control execution may a better way to increase the utilization of pre trained layers and models. Fostering a pattern in which multiple models are shared as different classes may minimize this problem. For example, publishing a TrainingRcnnModel and InferenceRcnnModel side by side with the same weights. I anticipate variable execution flows to become more common in the future as architectures increase in complexity.

PR #30, opened 2021-09-23
Fix minor typo on guidelines

ADGEfficiency:
Minor typo fix:

```
... confuse our users (like, say, `Network` and `Model` -- **since** is why we don't export `Network` as a public API).
```

Becomes:
```
... confuse our users (like, say, `Network` and `Model` -- **this** is why we don't export `Network` as a public API).
```


ADGEfficiency:
@googlebot I signed it!

haifeng-jin (approved):
LGTM

PR #31, opened 2021-12-06
project setup best practices

haifeng-jin:
This is an open discussion on how to set up the open-source projects under `keras-team/keras`.

### DIFF
--- a/project_setup_best_practices.md
+++ b/project_setup_best_practices.md
@@ -66,7 +66,22 @@ Fix the random seed for all tests:
 
 Create a temporary path for testing: [Link](https://docs.pytest.org/en/6.2.x/tmpdir.html).
 
-## Linting and formatting
+## Code styles
+
+### Importing Keras modules
+
+For projects based on Keras and TensorFlow, top-level imports are encouraged, like
+shows in the following example.
+
+```py
+import tensorflow as tf
+from tensorflow import keras
+```
+
+Exceptions may be acceptable when the module appeared too many times in the code,
+like `keras.layers`.
+
+### Linting and formatting
 
 We use
 [black](https://black.readthedocs.io/en/stable/),
--- a/project_setup_best_practices.md
+++ b/project_setup_best_practices.md
@@ -37,8 +37,8 @@ pattern of `*_test.py`. For example, the tests for
 `/keras_tuner/engine/hyperparameters.py` are in
 `/keras_tuner/engine/hyperparameters_tests.py`.
 
-Integration tests may be contained in their own `/integration` directory, as
-they may require extra files such as data.
+Integration tests may be contained in their own `/keras_tuner/integration_tests`
+directory, as they may require extra files such as data.
 
 While our unit test placement is not suggested in the
 [good practices of pytest](https://docs.pytest.org/en/6.2.x/goodpractices.html)
--- a/project_setup_best_practices.md
+++ b/project_setup_best_practices.md
@@ -1,0 +1,165 @@
+# Best Practices for Managing Keras Projects on GitHub
+
+This document describes the best practices for managing the projects under
+"keras-team" on GitHub which use GitHub as the source of truth, including
+[keras-tuner](https://github.com/keras-team/keras-tuner),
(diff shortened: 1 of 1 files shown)


project_setup_best_practices.md
```diff
@@ -0,0 +1,165 @@
...
+Fix the random seed for all tests:
+[Link1](https://github.com/keras-team/keras-tuner/blob/1.1.0/tests/conftest.py#L8-L17),
+[Link2](https://github.com/keras-team/keras-tuner/blob/master/tests/unit_tests/randomness_test.py),
+[Link3](https://www.tensorflow.org/api_docs/python/tf/keras/utils/set_random_seed).
+
+Create a temporary path for testing: [Link](https://docs.pytest.org/en/6.2.x/tmpdir.html).
+
+## Linting and formatting
+
+We use
```
bhack, line 86: The behavior of these tools could really change on the installed version. It is why is suggest to enforce the versioning with `pre-commit install` and configuration like at https://github.com/keras-team/keras-cv/pull/14.
I suggest to use the docker developer image and `pre-commit` also in  Githun Action CI so that we are sure we are always on the same page with the developers when they validate the in-progress PR code on their own env

project_setup_best_practices.md
```diff
@@ -0,0 +1,165 @@
...
+[black](https://black.readthedocs.io/en/stable/),
+[isort](https://pycqa.github.io/isort/), 
+[flake8](https://flake8.pycqa.org/en/latest/)
+to lint and format the code. black is to generally format the code. isort is to
+sort the imports. flake8 is for some additional checks that black doesn't do,
+like the long lines with a single string. You can see the relevant sections of
+[setup.cfg](https://github.com/keras-team/keras-tuner/blob/1.1.0/setup.cfg) for
+the detailed configuration of these tools.
+
+The user does not need to know how to use these tools to lint or format the
```
bhack, line 96: Generally users don't run script before submitting PR or update PR with new commits this is why is suggest to rely on commit hooks.
`precommit-hooks` proposed in https://github.com/keras-team/keras-cv/pull/14 could help to quickly install the right version of the tools and activate hooks.

bhack:
I've added a few comment. 

More in general I think that we need to distribute an official Docker devel image that could be used also in Github Action to be sure that developer and CI work always on the same page as by my experience there is a natural trend over time to accumulate a drift between local developer env and the CI.

Please check also if you could find something interesting in the proposal about standardizing Github community health files as proposed some months ago in:
https://discuss.tensorflow.org/t/standardize-github-community-health-files-in-the-ecosystem/1708

project_setup_best_practices.md
LukeWood: Hey @bhack I have some questions regarding the security of pre-commit.  Do we know how the tool is versioned/etc?  I'm a little uneasy locally running code we pull from a remote repo - though if it is inside of a docker container it should be fine.

LukeWood:
I think @bhack has some good points on standardization of pre-commits with the pre-commit package.  Let's discuss next week

project_setup_best_practices.md
```diff
@@ -0,0 +1,180 @@
...
+code. We provide them with two shell scripts:
+[`/shell/lint.sh`](https://github.com/keras-team/keras-tuner/blob/master/shell/lint.sh)
+and
+[`/shell/format.sh`](https://github.com/keras-team/keras-tuner/blob/master/shell/format.sh).
+In these scripts, we also check and add the Apache 2.0 License head to every
+file.
+
```
bhack, line 103: Are these scripts globals for the Keras ecosystem or do we have a baseline template that need to be customized for every repository? Also if we are going to customize this scripts for every repository they will go naturally out of sync over time if we don't handle a master.

We partially discussed this mentioning https://stackoverflow.com/questions/59047996/how-can-i-configure-pre-commit-tool-to-rely-on-a-single-location-for-all-hook-co

But as currently we don't want to have automatic pre-commit hooks checks before commits we could probably evaluate to set a specif repository collecting `master templates` of these common infra scripts.

fchollet (approved):

PR #35, opened 2022-02-12
Fix minor typo on keras_api_design_guidelines.md

hsjplus:
I use Keras a lot for ML projects and now seek if I can contribute to Keras. So I was reading keras_api_design_guidelines.md, and found a minor typo. It would be more proper to use "a very poor error message" instead of "an very poor error message".


fchollet (approved):
LGTM, thanks

